PR #13, opened 2024-04-21
Fix mypy

maresb:
It seems that mypy broke for some reason after the migration. :man_facepalming:


tomers (approved):

maresb:
Ya, these issues should be addressed, but I don't want to create a merge conflict with #12. How would you like to proceed?

tomers:
I'll rebase on top of this, and fix the error in my PR. Thanks for suggesting this.

maresb:
Rebased into #12

PR #15, opened 2024-04-21
Remove vscode settings file

maresb:
I think including a launch.json is already a bit borderline, but we really shouldn't have a settings.json.

In order to get the ruler in the dev container, I added it to the vscode customizations.

Were those other settings intended to fix any particular breakage?


tomers (approved):
LGTM!

PR #20, opened 2024-04-25
A few more device selector fixes

maresb:
We aren't testing the GUI in the automated tests, and I discovered it was failing to load on Python 3.8 due to missing future annotations. Ideally this stuff is caught by tests, but I've never done a GUI before, so I don't know offhand how to test a GUI from GitHub CI.

Next, having a greedy `--device` argument was conflicting with the greedy `text` argument. I changed it so that you add multiple filters via multiple `--device` arguments.

Does it make sense to be matching on the manufacturer? I'm inclined to just check product description and serial number.

Finally, I'm not sure offhand how to solve this, but let's say I have multiple devices and I want to determine a filter string. Currently the only way I can see to do this is to print a label and include `--verbose`, which I find pretty unintuitive. Perhaps we need something like `--list-devices`, but I feel like our arguments are getting fairly bloated.

### DIFF
--- a/src/labelle/cli/cli.py
+++ b/src/labelle/cli/cli.py
@@ -202,8 +202,11 @@ def parse_args():
     )
     parser.add_argument(
         "--device",
-        nargs="+",
-        help="Device pattern to match",
+        action="append",
+        help=(
+            "Select a particular device by filtering for a given substring "
+            "in the device's manufacturer, product or serial number."
+        ),
         type=str,
     )
     return parser.parse_args()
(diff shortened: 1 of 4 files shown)


tomers:
I already made some improvements to the CLI, so I had to manually repeat your changes.
I already removed the eagerness of the text label, so I hope there should not be any issue with how it is currently.
I plan to implement `--list-devices` soon.

tomers:
Please add task to add automated tests for GUI :pray:

maresb:
> Please add task to add automated tests for GUI 🙏

@tomers, I wish I knew how, I'll have to look into it.

I guess one really quick test that would have caught these things would be to import all modules.

PR #23, opened 2024-04-27
Fix barcodes in GUI

maresb:


src/labelle/lib/render_engines/barcode.py
```diff
@@ -25,6 +25,11 @@ def __init__(self, content: str, barcode_type: str | None):
         self.barcode_type = barcode_type or DEFAULT_BARCODE_TYPE
 
     def render(self, context: RenderContext) -> Image.Image:
+        if self.barcode_type == "code128" and self.content == "":
```
tomers, line 28: I suggest you use the global constant to default type, I stead of the string value, as it might change in the future

PR #25, opened 2024-04-27
Clean up barcodes

maresb:
This should be a case study for how _not_ to do object-oriented programming.

It's as if someone took a perfectly reasonable function and asked, "Ok, how can we maximally obfuscate this so that nobody can figure out how it works? Oh, I know, we can add a bunch of unnecessary global state, and make the control flow nondeterministic."


maresb:
(No content changes, just a fresh rebase on `main`.)

maresb:
Let's merge #32 first. I checked these changes by hand, but it'll be great to be sure I didn't miss anything.

PR #35, opened 2024-04-28
Add deprecation errors for old dymoprint arguments

maresb:


tomers (approved):
I wonder how many users do we currently have. Answer to this should determine whether we should invest in such efforts, and keep so much code related to transition to the new interface.

I got the impression that since our device support is so narrow at the moment, we don't have large user base, so we can move fast and break things.
Despite that, I appreciate your professionalism of making sure our existing users get a smooth experience transitioning to the latest code.

maresb:
You're right it's probably not worth the effort, but I was thinking of tweaking the deprecation message in dymoprint to mention that Labelle v1 is a drop-in replacement for dymoprint.

maresb:
> Let's go for it then, but I would like to remove these legacy entries when we release a new major version with breaking changes.

Absolutely. v2 should be a clean break. We can simply revert this commit. This was also somewhat of a sanity check to ensure that we're not missing anything, and indeed it uncovered #33 and #34.

PR #37, opened 2024-04-30
GUI tests

tomers:
The GUI is rather useless when the app is being run in the test environment, as there are no printers connected (there is no USB backend at all).
This simple test verifies the most basic existence of selected widget, and their content.
We will need to figure out how to test the app with a state that resembles actual printer. We might need to introduce test mode, in which all supported printers are shown in the device selector, so we can test various features for each.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,8 @@ jobs:
         os : [ubuntu-latest]
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
       fail-fast: false
-
+    env:
+      DISPLAY: ':99.0'
     steps:
     - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


tomers:
@maresb please note that this branch is based on the `tshalev-add-tests` branch.
Actually, this PR contains only a single commit (at the moment, that might change later on).

maresb:
@tomers, any ideas what's going wrong here? It appears that the Python process inside tox is crashing for some reason.

tomers:
> @tomers, any ideas what's going wrong here? It appears that the Python process inside tox is crashing for some reason.

I suspect this: https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#tox-invocationerror-without-further-information
> `pytest-qt` needs a `DISPLAY` to run, otherwise `Qt` calls `abort()` and the process crashes immediately.

tomers:
I will follow instructions in https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#github-actions-azure-pipelines-travis-ci-and-gitlab-ci-cd

maresb:
Nice. Also, we don't need to be using tox. It may be less trouble to switch to raw `pytest` and use a GitHub Actions matrix to check the various Python versions.

tomers:
@maresb I think this PR is ready to be merged :-)

maresb (approved):

maresb:
Woah, great work!!!

Go ahead and merge, but for future reference could you please add a comment here about the references you used for:
1. the Ubuntu dependencies
2. the start-stop-daemon command?

I don't know where they came from, and they may need to be adjusted in the future, so I'd just like to gather the info while it's still fresh in your head.

maresb:
I'm really impressed, this looked like a hard one, so I wasn't expecting it to get done today. Very nice!!!

tomers:
> For future reference could you please add a comment here about the references you used for:

The comment in the second commit gives the proper reference. I've committed some changes in other commit by mistake, and now I've arranged it properly, so all dependency handling is done in a single commit, with proper documentation.

Thanks for your kind words!

PR #52, opened 2024-06-17
Update dev environment

FaBjE:
I've made some changes to the dev container/environment Two small, one big.

Small changes:
- Replaced deprecated command line option with "modern" one.
- Add extra debug launch config for printer (see big change)

Big change:
- Allow using USB printers from the devcontainer. You can now print directly from the docker container to the dymo (On Linux host devices)

let me know what you think 😃

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,6 +9,8 @@ ARG deps="\
     libxcb-keysyms1 \
     libxcb-shape0 \
     libxkbcommon-x11-0 \
+    libusb-1.0-0 \
+    usbutils \
     "
 RUN apt-get update \
     && apt-get install -y ${deps} \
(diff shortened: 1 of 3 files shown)


maresb:
This is really amazing work @FaBjE, thanks so much for the contribution!!!

FaBjE:
@maresb Did you test your fix? I'm not sure (and cant test it atm), but I think the argument needs to be quoted on the commandline (when vs code is starting docker) because it contains spaces. The outer quotes are for the JSON. the single `'` inner quotes were for the commandline.

maresb:
Ya, I tested it. My fix is consistent with the link you provided in combination with the way shells process arguments. When you write `'c 189:* rw'` in a shell, the shell parses it so that it is a single argument rather than `c`, `189:*`, `rw`. When specifying arguments as a JSON list, each list element becomes a single unquoted argument. Before my fix, the quotation marks were interpreted as part of the argument itself, not as grouping characters.

Shells (especially Bash) are weird.

FaBjE:
Ah good to know! I always get confused with shells, quotation and especially when providing the input via a 3rd party like the JSON. 
Oddly enough it worked for me with the quotes, but It's always better to do it the proper way 😃

maresb:
Strange, it was broken for me before the fix.

PR #11, opened 2024-04-15
Tomer's USB refactor

tomers:
Some initial work towards supporting labeler device selector widget.
@maresb, please take extra care reviewing this, as I had to duplicate my changes in the new repository. I hope I did not revert any of your recent changes related to rebranding the software.

PR #54, opened 2024-06-21
Device dependent margins/config

### LINKED ISSUE #51: Adding "DYMO LabelMANAGER PC II" support (help requested)
I was trying to see if I can get my DYMO LabelMANAGER PC II to work with Labelle.
I saw that the "DYMO LabelMANAGER PC" is supported, so I gave it a shot.

It is not supported out of the box, but when adding the following line to constants.py:
`0x001C: "DYMO LabelMANAGER PC II",`
and setting up the udev rules, the printer is detected and I can print a label!

But... 

The text is not aligned correctly. I think this is because it is a 24mm tape printer.
I did some digging around and found these lines in dymo_labeler.py
```
    LABELER_DISTANCE_BETWEEN_PRINT_HEAD_AND_CUTTER_MM = 8.1
    LABELER_PRINT_HEAD_HEIGHT_MM = 8.2
    SUPPORTED_TAPE_SIZES_MM = (19, 12, 9, 6)
    DEFAULT_TAPE_SIZE_MM = 12`
```

I added `SUPPORTED_TAPE_SIZES_MM = (24, 19, 12, 9, 6)` and can print a correctly aligned 24mm label!
But the 12mm sizes are still not aligned correctly.
Changing LABELER_PRINT_HEAD_HEIGHT_MM to 24 did not fix it.

My questions:
- How do I align the label of the smaller tape sizes?
- How do I add these changes without breaking other printers?

My python knowledge is next to nothing, so patient please :)

FaBjE:
Closes #51

I have added a DeviceConfig object that contains a configuration for each type of printer/labeler.
I started out with the lookup table, but puzzeled by the "random" margins I figured it out how the principal works.

For every printer we now need a "print head size in pixels" and "print head active size in mm" the rest is calculated.
I think there might only be two versions (128px/18mm and 64px/9mm).  
Based on my scans we determined a label-position inaccuracy of 0.7mm. I rounded it up to 1mm to use as safety margin for now.

I´d like to hear your thoughts. Bear in mind that this is my first python code. So you may find a lot of weird things...
This is not done yet, so only a draft PR.

I've tested it on my LabelManager PC II and a  LabelManager PnP both can print valid labels 12mm (and 24mm on the PCII)

I have identified a couple of issues so far:
- Supported tape sizes dropdown not cleared in GUI, making a mess. With the config object it should now show only the options supported by the printer
- Margins preview in GUI are not correct
- Boxed text, left border 1px missing
- mm_to_px functions are maybe not accurate. Look into/refactor those?


src/labelle/gui/q_settings_toolbar.py
```diff
@@ -57,7 +57,7 @@ def _init_elements(self) -> None:
 
     def update_labeler_context(
         self,
-        supported_tape_sizes: tuple[int, ...],
+        supported_tape_sizes: list[int],
```
tomers, line 60: Is this really needed? In case the list is constant, it would be better as a convention to prefer tuple over list

src/labelle/lib/constants.py
```diff
@@ -34,20 +34,56 @@
...
+    LABELMANAGER_420P_MODE_SWITCH = 0x1004
+    LABELMANAGER_280P_NO_MODE_SWITCH = 0x1005
+    LABELMANAGER_280P_MODE_SWITCH = 0x1006
+    LABELMANAGER_WIRELESS_PNP_NO_MODE_SWITCH = 0x1007
+    LABELMANAGER_WIRELESS_PNP_MODE_SWITCH = 0x1008
+    MOBILE_LABELER = 0x1009
+
+
+# fmt: off
+# Very bad I know, but it keeps a loop of removing the line breaks and errroring
```
tomers: Try to use slash at the end of the line that you break, .e.g.:
```
    SUPPORTED_DEVICE_ID.LABELMANAGER_PC: \
        "DYMO LabelMANAGER PC",
```

src/labelle/lib/devices/dymo_labeler.py
```diff
@@ -240,23 +241,81 @@ class DymoLabeler:
...
 
+    # Default supported tape sizes
+    labeler_supported_tape_sizes: list
+
+    # Default printer head size in dots
+    labeler_print_head_width_pixels: int
+
+    # Default printer margins (for each tape size)
+    # { tapeSizeMm : (firstVisiblePixel, lastVisiblePixel) }
+    labeler_label_vertical_margins: dict[int, Any]
```
tomers: Maybe something like:
```
labeler_label_vertical_margins: dict[int, tuple[int, int]]
```

src/labelle/lib/devices/dymo_labeler.py
```diff
@@ -240,23 +241,81 @@ class DymoLabeler:
...
+
     def __init__(
         self,
         tape_size_mm: int | None = None,
         device: UsbDevice | None = None,
     ):
         if tape_size_mm is None:
             tape_size_mm = self.DEFAULT_TAPE_SIZE_MM
```
tomers, line 255: Is this still relevant? Is `DEFAULT_TAPE_SIZE_MM` supported by all printer models? I guess we should choose the first available tape size for the model instead.

src/labelle/gui/gui.py
```diff
@@ -63,7 +63,7 @@ def _init_elements(self) -> None:
         self._device_manager = DeviceManager()
         self._dymo_labeler = DymoLabeler()
         self._settings_toolbar.update_labeler_context(
-            supported_tape_sizes=self._dymo_labeler.labeler_supported_tape_sizes,
+            supported_tape_sizes=self._dymo_labeler._deviceConfig.supportedTapeSizes,
```
tomers, line 66: Maybe instead of exposing inner `_deviceConfig`, add property to the labeler class to which exposes it.

src/labelle/lib/constants.py
```diff
@@ -52,36 +53,110 @@ class SUPPORTED_DEVICE_ID(Enum):
...
 
-# fmt: off
-# Very bad I know, but it keeps a loop of removing the line breaks and errroring
+# Create supported products list
+SUPPORTED_PRODUCTS = []
 
-SUPPORTED_PRODUCTS = {
-    SUPPORTED_DEVICE_ID.LABELMANAGER_PC:
+# ---- Supported USB Devices configuration ----
+SUPPORTED_PRODUCTS.append(
```
tomers, line 60: I think that eventually we should move these out of the code, into a JSON file. If possible, please try to do it. In not, we can try refactor it later.

src/labelle/lib/constants.py
```diff
@@ -52,36 +53,110 @@ class SUPPORTED_DEVICE_ID(Enum):
...
+    DeviceConfig(
         "DYMO LabelMANAGER PC",
-    SUPPORTED_DEVICE_ID.LABELPOINT_350:
+        [int(SUPPORTED_DEVICE_ID.LABELMANAGER_PC)],
+        # ToDo: Validate config!
+        128,
```
tomers, line 66: It's hard to understand from reading this what each field means. Please explicitly name each field.

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+    """Size of the print head in pixels (use calibration routine to determine)"""
+
+    supportedTapeSizes: list[int]
+    """List of supported tape sizes in mm"""
+
+    marginsPerTape: dict[int, Tuple[int, int]]
+    """
+    Dictonary of print margins per tape size
+    Entry: TapeSizeMM : (topMarginInPx, bottomMarginInPx)
+    Use the calibration routine to determine these for each tape size
```
tomers: Please add a link here to a markdown document under `/doc` which describes what the calibration routine is.

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    def __init__(
+        self,
+        name: str,
+        deviceIDs: list[int],
```
tomers, line 39: Please avoid using camelCase, and instead use smake_case. For example `device_ids`, not `deviceIDs`.

https://stackoverflow.com/questions/159720/what-is-the-naming-convention-in-python-for-variables-and-functions

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+        printHeadSizePixels: int,
+        supportedTapeSizes: list[int],
+        marginsPerTape: dict[int, Tuple[int, int]],
+    ):
+        """Initialize a Labeler config object."""
+        self.name = name
+        self.deviceIDs = deviceIDs
+        self.printHeadSizePixels = printHeadSizePixels
+        self.supportedTapeSizes = supportedTapeSizes
+        self.marginsPerTape = marginsPerTape
```
tomers: Maybe define the class as `@dataclass`. https://docs.python.org/3/library/dataclasses.html

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def matches_device_id(self, idValue: int) -> bool:
+        """Check if the a device ID matches this config."""
+        if idValue in self.deviceIDs:
+            return True
+        else:
+            return False
```
tomers, line 56: I am not sure a function is really needed for this. In any case, it should be refactor as follows:
```
return idValue in self.deviceIDs
```

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def get_tape_print_margins(self, tapeSizeMM: int) -> tuple[int, int]:
```
tomers: Also here, `tape_size_mm`, not `tapeSizeMM` (I will refrain now from commenting of improper casing).

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+        """Get print margins for a specific tape size.
+
+        :param tapeSizeMM: Tape size in mm
+        :return: Margins tuple, None if not supported
+        """
+        if tapeSizeMM in self.supportedTapeSizes:
```
tomers, line 66: As coding convention, instead of:
```
    if term:
        # some very long logic
    else:
        raise ...
```
do:
```
    if term:
        raise ...
    # some very long logic
```
This reduces nesting.

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+                return self.marginsPerTape[tapeSizeMM]
+            else:
+                # No specific margins set, return default
+                return (0, 0)
+        else:
+            # Tape size not supported
+            raise ValueError(
+                f"Unsupported tape size {tapeSizeMM}mm. "
+                f"Supported sizes: {self.supportedTapeSizes}mm"
+            )
```
tomers: As coding convention, instead of:
```
    if term:
        # some very long logic
    else:
        raise ...
```
do:
```
    if term:
        raise ...
    # some very long logic
```
This reduces nesting.

src/labelle/lib/devices/device_config.py
```diff
@@ -0,0 +1,71 @@
...
+        :param tapeSizeMM: Tape size in mm
+        :return: Margins tuple, None if not supported
+        """
+        if tapeSizeMM in self.supportedTapeSizes:
+            if tapeSizeMM in self.marginsPerTape:
+                # Return specified margins
+                return self.marginsPerTape[tapeSizeMM]
+            else:
+                # No specific margins set, return default
+                return (0, 0)
```
tomers:
```
return self.marginsPerTape.get(tapeSizeMM, (0, 0))
```
or even better:
```
DEFAULT_MARGINS = (0, 0)  # put this on top of file
return self.marginsPerTape.get(tapeSizeMM, DEFAULT_MARGINS)
```

src/labelle/lib/devices/device_manager.py
```diff
@@ -87,3 +88,18 @@ def find_and_select_device(self, patterns: list[str] | None = None) -> UsbDevice
...
+
+def get_device_config_by_id(idValue: int) -> DeviceConfig:
+    """Get a labeler device config with USB ID.
+
+    :param idValue: USB ID value
+    :return: Device config, None if not found
+    """
+    #
+    for device in SUPPORTED_PRODUCTS:
+        if device.matches_device_id(idValue) is True:
```
tomers, line 103: Remove `is True`. It will work the same

src/labelle/lib/devices/dymo_labeler.py
```diff
@@ -254,14 +254,15 @@ def __init__(
 
         if self._device is not None:
             # Retrieve device config
-            self._deviceConfig = get_device_config_by_id(self._device.id_product)
+            foundConfig = get_device_config_by_id(self._device.id_product)
+            if foundConfig is not None:
+                self._deviceConfig = foundConfig
+            else:
+                raise ValueError(f"Unsupported device type {self._device.id_product}")
```
tomers: As before, style comment. Please use the following convention:
```
            if foundConfig is None:
                raise ValueError(f"Unsupported device type {self._device.id_product}")
            self._deviceConfig = foundConfig

```

tomers: Why do you return `None` from the function and raise here, instead of raising in the function itself, and then you will not have to test returned value from the function (and change its type to also support `None`).

src/labelle/lib/devices/device_manager.py
```diff
@@ -82,8 +81,11 @@ def find_and_select_device(self, patterns: list[str] | None = None) -> UsbDevice
             LOG.debug(dev.device_info)
         dev = devices[0]
         if dev.is_supported:
-            msg = f"Recognized device as \
-                {SUPPORTED_PRODUCTS[SUPPORTED_DEVICE_ID(dev.id_product)]}"
+            foundDeviceConfig: DeviceConfig | None = get_device_config_by_id(
```
tomers, line 84: Do you really need to declare the type of `foundDeviceConfig` (I guess you meant `found_device_config` :-) )?
If the function declares its return type, then I think it should be fine not to declare here.

src/labelle/lib/devices/device_manager.py
```diff
@@ -82,8 +81,11 @@ def find_and_select_device(self, patterns: list[str] | None = None) -> UsbDevice
...
+                dev.id_product
+            )
+            if foundDeviceConfig is not None:
+                msg = f"Recognized device as {foundDeviceConfig.name}"
```
tomers, line 88: What is `msg` if `found_device_config` is `None`?

src/labelle/lib/devices/device_config.py
```diff
@@ -18,30 +18,35 @@ class DeviceConfig:
...
     """List of supported tape sizes in mm"""
 
-    marginsPerTape: dict[int, Tuple[int, int]]
-    """
-    Dictonary of print margins per tape size
-    Entry: TapeSizeMM : (topMarginInPx, bottomMarginInPx)
-    Use the calibration routine to determine these for each tape size
-    """
+    # Fixed to 1 mm until proven otherwise ;)
+    TAPE_ALIGNMENT_INACCURACY_MM: float = 1
```
tomers, line 29: Since it is float, please write `1.0`.

src/labelle/lib/devices/device_config.py
```diff
@@ -18,30 +18,35 @@ class DeviceConfig:
...
+    """The inaccuracy of the tape position relative to the printhead"""
+    # Inaccuracy of the tape position is mostly caused by
+    # the tape moving slightly from side to side in the cartridge.
+    # Improper cartrigde placemement is also an factor,
+    # but negligible due to a physical spring in the lid.
```
tomers, line 34: Please make lines longer.
I think you should be able to write all comments under a single `""" .... """` block, just make sure to add the extended description following a separating empty line. https://stackoverflow.com/a/23188939/1692287

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
...
+            # Calculate the pixels per mm for this printer
+            # Example: printhead of 128 Pixels, distributed over 18 mm of active area.
+            #   Makes 7.11 pixels/mm
+            printPPmm: float = (
+                self.printHeadSizePixels / self.printHeadActiveAreaMillimeters
+            )
+
+            # Calculate usable tape width (*2 for top and bottom)
+            usableTapeWidth: float = tapeSizeMM - (
+                self.TAPE_ALIGNMENT_INACCURACY_MM * 2
```
tomers, line 76: I guess this is a really silly comment, but since in my mind I read "tape width minus twice the margins", I would write `2 * self.TAPE_ALIGNMENT_INACCURACY_MM` instead of `self.TAPE_ALIGNMENT_INACCURACY_MM * 2`

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
...
-                return self.marginsPerTape[tapeSizeMM]
+            # Calculate the pixels per mm for this printer
+            # Example: printhead of 128 Pixels, distributed over 18 mm of active area.
+            #   Makes 7.11 pixels/mm
+            printPPmm: float = (
+                self.printHeadSizePixels / self.printHeadActiveAreaMillimeters
+            )
+
+            # Calculate usable tape width (*2 for top and bottom)
+            usableTapeWidth: float = tapeSizeMM - (
```
tomers, line 75: Please make this a property, not a calculated variable inside a function.

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
         else:
             return False
 
-    def get_tape_print_margins(self, tapeSizeMM: int) -> tuple[int, int]:
+    def get_tape_print_size_and_margins_px(
```
tomers, line 58: Please make this a property, rather a `get_` function.

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
...
+            activeTapePixels = math.floor(activeTapePixels)
+
+            # To calculate the margins we need to know some hardware info
+            # Printer has special "support studs" that
+            # let 19 & 24mm tapes go to the bottom (if supported)
+            # but 12mm based casettes are raised so they are centered to the printhead.
+            # Smaller tapes than 12mm are in 12mm casettes are centered
+            # in the cartridge and in turn also centered to the printhead
+            # Which gives us the advantage that we can
+            # just calculate the top and bottom margin
```
tomers, line 99: Please break lines according to our linter line width (120 chars). Having too short lines makes it harder to read.

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
...
+                activeTapePixels = printPPmm * usableTapeWidth
+
+            # Round down to nearest whole number as we can't use half a pixels ;)
+            activeTapePixels = math.floor(activeTapePixels)
+
+            # To calculate the margins we need to know some hardware info
+            # Printer has special "support studs" that
+            # let 19 & 24mm tapes go to the bottom (if supported)
+            # but 12mm based casettes are raised so they are centered to the printhead.
+            # Smaller tapes than 12mm are in 12mm casettes are centered
```
tomers, line 96: I think some rephrasing is needed here: "are in 12mm casettes are centered"

src/labelle/lib/devices/device_config.py
```diff
@@ -50,19 +55,57 @@ def matches_device_id(self, idValue: int) -> bool:
...
+            topMargin = round(((self.printHeadSizePixels - activeTapePixels) / 2), 0)
+
+            # Bottom margin is equal due to centering of the tape
+            bottomMargin = topMargin
+
+            # Return active pixels / margins set
+            return (int(activeTapePixels), int(topMargin), int(bottomMargin))
         else:
             # Tape size not supported
             raise ValueError(
```
tomers: Here we encounter handling of branch from 60 lines ago... so instead, why not in line 66:
```
if tapeSizeMM not in self.supportedTapeSizes:
    raise ...
```

BTW I know this is not your code, but I comment on everything that I notice while reading your changes, which also include existing code. I hope you are OK with that.

src/labelle/lib/devices/dymo_labeler.py
```diff
@@ -274,7 +273,10 @@ def __init__(
 
     @property
     def height_px(self):
-        return DymoLabelerFunctions.height_px(self.tape_size_mm)
+        """Get the tape size in pixels."""
+        return self._deviceConfig.get_tape_print_size_and_margins_px(self.tape_size_mm)[
+            0
+        ]
```
tomers, line 268: Using `[0]` is rather odd. Please refer to my earlier comment about using properties.

src/labelle/lib/devices/dymo_labeler.py
```diff
@@ -312,6 +301,17 @@ def device(self, device: UsbDevice | None):
         try:
             if device:
                 device.setup()
+
+                # Retrieve device config
+                foundConfig = get_device_config_by_id(device.id_product)
```
tomers, line 306: Please raise inside `get_device_config_by_id`

tomers (commented):
Thanks for this, @FaBjE ! This is great stuff. Your contribution is important, since label size handling was a bit shady until now.
Please refer to my comments. Most of them are coding style comments.
If possible, try to rebase fixes onto existing commits, to me review easier on us.
Please review commit messages, are some seems unrelated to the content of the commit.
Again, thank you for your important work!
I hope you don't mind to do a round of two of fixing review comments before your work could be integration into the codebase.

src/labelle/lib/devices/device_config.py
tomers: Actually, it is 88 chars

src/labelle/lib/devices/dymo_labeler.py
maresb: I am wondering what "first" and "last" mean here regarding visible pixels. I don't know offhand which pixel ordering is being used, so it would be helpful to be more explicit.

src/labelle/lib/constants.py
maresb: I think we need a proper data model(1) for device models(2). (The first "model" in that sentence means "implementation of a schema" while the second "model" refers to a particular product that is sold by a company.)

src/labelle/lib/devices/device_config.py
maresb: Personally, I'd probably use a [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) which is simpler, but either should work.

src/labelle/lib/constants.py
maresb: Sorry, I was a bit confused You do already define `DeviceConfig`.

I'm wondering how to achieve this elegantly. I agree that it's in principle better to define these things out-of-code like in JSON. However, that unfortunately would require deserialization. Pydantic is a popular library for this, but then we're adding more project dependencies and a lot of complexity, also in the code.

The way I see it, we have two main options:
1. Separate the device definition from code, perform deserialization, probably add a dependency on something like Pydantic.
2. Keep the device definitions as code, but switch to `NamedTuple` or `dataclass`.

Another disadvantage to 1. is that editing the config would involve editing serialized data which can be a real pain. An advantage to 2. is that code tools like mypy will automatically perform validation for the data-as-code.

I'm leaning more towards 2.

src/labelle/lib/constants.py
```diff
@@ -34,19 +35,138 @@
...
+    LABELMANAGER_420P_MODE_SWITCH = 0x1004
+    LABELMANAGER_280P_NO_MODE_SWITCH = 0x1005
+    LABELMANAGER_280P_MODE_SWITCH = 0x1006
+    LABELMANAGER_WIRELESS_PNP_NO_MODE_SWITCH = 0x1007
+    LABELMANAGER_WIRELESS_PNP_MODE_SWITCH = 0x1008
+    MOBILE_LABELER = 0x1009
+
+
+# Create supported products list
+SUPPORTED_PRODUCTS = []
```
maresb, line 57: Using `append` is a dynamic operation. I think it would look better to do

```python
SUPPORTED_PRODUCTS = [
    DeviceConfig(
        ...
```

maresb:
Thanks @FaBjE for the awesome contribution and @tomers for such a detailed review! I'm pretty excited about finally getting this feature settled.

I agree that we'll want to do another round of review after addressing these structural and stylistic issues.

> If possible, try to rebase fixes onto existing commits, to me review easier on us.

I'm wondering what is exactly your preference here @tomers. Most reviewers prefer to avoid rebasing until the end of the review process so that there is a clear history of the changes. However, GitHub introduced a feature that makes it easy to diff before and after a rebase, so this is no longer such a big deal. I don't have such a strong preference, but I'd like to understand your rationale. (For tiny PRs I slightly prefer rebasing, but with larger PRs like this one I like new commits.)

tomek-szczesny:
(Comment copied from the issue)

Hm. The point of LUT that I developed was to be prepared for any future Dymo printer. That is why I proposed to measure print area in dots, as well as a single necessary offset, and then only name this settings in mm for user's convenience.
No millimeters handled in the code because frankly we don't need them there at all.

I'm afraid that although your solution is convenient for supporting a subset of Dymo products known to us at the moment, I'm not sure how well this will scale with printers of other brands or series.

tomers:
> I'm wondering what is exactly your preference here @tomers. Most reviewers prefer to avoid rebasing until the end of the review process so that there is a clear history of the changes. However, GitHub introduced a feature that makes it easy to diff before and after a rebase, so this is no longer such a big deal. I don't have such a strong preference, but I'd like to understand your rationale. (For tiny PRs I slightly prefer rebasing, but with larger PRs like this one I like new commits.)

I am not orthodox on this, so I guess I can live with whatever approach used. I guess it is all about the convenience of the author and reviewers. If the author doesn't feel comfortable with rebasing, let him do whatever he wants, as to not put a burden on him. I am not familiar with Github tooling that much. I personally don't like the review system, as it is not straight forward to my opinion, and makes it hard to track progress. In my work we use https://rbcommons.com/ which is a good tool to monitor changes to a given commit.

As you saw in the past, I personally prefer to keep contained list of commits, and keep rebasing the same commits over and over, until the change is stable, and each commit containes a specific subset of the total work. In this approach you will not encounter TODO section, and then a removal of that TODO in a subsequent commit.

I agree that the approach depends on the size of the given review, and the number of iterations in its lifetime. Using the rebase approach indeed makes it harder for reviewers to track changes in the review.

So I'll leave it for @FaBjE to decide how he would like to work on this review. I want him to enjoy his work, and keep his motivation to contribute to the project :-)

src/labelle/lib/constants.py
maresb: To be explicit, this is as simple as

```python
DeviceConfig(
    name="DYMO LabelMANAGER PC",
    deviceIDs=[int(SUPPORTED_DEVICE_ID.LABELMANAGER_PC)]
    ...
```

Also, the use here of `SUPPORTED_DEVICE_ID` seems very indirect to me. Why not just do

```
    deviceIDs=[0x0011]
```

FaBjE:
@All Thank you for your feedback guys
I completely understand a lot of style issues. I'm not happy with most of them. But it are either issues of me not knowing python syntax well enough or the linter/validator settings not allowing me to.
let me address your points:

> Hm. The point of LUT that I developed was to be prepared for any future Dymo printer. That is why I proposed to measure print area in dots, as well as a single necessary offset, and then only name this settings in mm for user's convenience. No millimeters handled in the code because frankly we don't need them there at all.
> 
> I'm afraid that although your solution is convenient for supporting a subset of Dymo products known to us at the moment, I'm not sure how well this will scale with printers of other brands or series.

I get your point and to emphasize it. I started out with exactly your solution. But I got stuck and it made me rethink it.
I think I have settled in "middle ground" but let me explain:
In the software there are two units mixed at the moment. Pixels (dots) and millimeters.
There is a "pixels per millimeter" constant somewhere. But I doubt it's validity.

The thing is the tape sizes are only known in millimeters. You could make this in pixels. But than I wondered, how many pixels is a particular tape size. Well, that depends on the printer. (thinking about the future) it depends on the pixels-per-millimeter of the print head. So I need to know the amount of pixels in the print-head.
Than my calculations didn´t workout. Why? Because the print-head pixels don´t match the largest tape-size it can print.
The PC II printhead active area is only 18mm as tested with your test-pattern. (I feel slightly scammed with my 24mm labels now, as they will never be printed full-size 🤔 ) 
So I added that configuration parameter as well.

I was able to directly determine tape size (in pixels) now. and I started working on the offets LUT. 
But not having any 6mm tape for example, I had to calculate them. and than it occurred to me that I had all the input parameters of my calculation in software. So why not [add the calculation there](https://github.com/labelle-org/labelle/pull/54/files#diff-755c01ca55978ef53bd17a5a9ececa9d8862d1b69a8a52768aae153970b12012R102). It still uses the tuple group of pixels offsets there.
I was thinking that if we encounter non-centered printers in the future we can add a (subset) of the LUT to the device configuration object. 
and make a "if not in LUT use calculation, otherwise use set value" kind of structure. Thinking we have the best of both worlds.

For my two test printers the calculation worked fine. So I left off at that point.

My expectation is that 90% of the printers will be valid, but there is always a exception.
An other "augment" is that I want to make adding printers as easy as possible. The LUT is a bit harder to generate for an inexperienced programmer than just measuring the size of your test-pattern print.

I hope you can agree with me, but I'm open to your thoughts.

> So I'll leave it for @FaBjE to decide how he would like to work on this review. I want him to enjoy his work, and keep his motivation to contribute to the project :-)

I'm not really sure what you are on to about rebasing. I'm used to just commiting all the changes. At the end I should do some squashing as I made a pretty mess by now. but I leave that till later.
In case other features are merged to master I rebase this featurebranch on master to be up to date again.
Does that work for you?

src/labelle/gui/q_settings_toolbar.py
FaBjE: The amount of supported tape sizes varies from printer to printer. I got an error that I tried to change the "tuple" size. A quick google search gave this as a solution. If there is a better solution please elaborate.

src/labelle/lib/constants.py
FaBjE: I though I removed this? I can´t find it at my current ¨state"

src/labelle/lib/devices/dymo_labeler.py
FaBjE: @tomers Thanks, will try again. Had some linter issues with this...

@maresb Me neither (at the time) Now I know it is "first visible pixel from the top" I will extend the comment.

src/labelle/lib/devices/dymo_labeler.py
FaBjE: I'm not really happy about this either, but the whole software kind of relies on it (also with the GUI)
Maybe I should convert it to take the smallest or largest supported tape size?

src/labelle/gui/gui.py
FaBjE: You mean with the `@property` thing? I'll give it a shot (sorry python noob here)

src/labelle/lib/constants.py
FaBjE: What I was trying to achieve here is a what in C would be "array of structs" for the "const" configuration.
If you have any suggestions how to do this elegantly in python, let me know...

My idea was that the (existing) labeler class took this as input to know the device's exact properties.

I agree that having this out of code would be nice. But can we please keep this contained?
I started out with "just adding my printer", and it is getting a bit out of hand already imho.

src/labelle/lib/constants.py
FaBjE: If that is possible can you give me a syntax example? (It would definitely be a lot clearer)

src/labelle/lib/devices/device_config.py
FaBjE: Will check

src/labelle/lib/devices/device_config.py
FaBjE: Sorry. If you check the link you've sent you know what my default programming language is 😉 
I'll to a cleanup/refactor before it gets merged.

src/labelle/lib/devices/device_config.py
FaBjE: I'm not sure of the pro's and con's of both. But I think dataclass is easier to convert to?
I'll give it a shot

src/labelle/lib/devices/device_config.py
FaBjE: Will fix. It seemed more "contained" to me than just accessing the array itself from outside the object.

src/labelle/lib/constants.py
maresb: @FaBjE that was an explicit example, and should work without other modifications to your code if you continue the obvious pattern. (Python lets you specify the names of positional arguments in the function invocation.) Does this make sense?

src/labelle/lib/devices/device_config.py
FaBjE: Please do 😉  I'll do a re-check of my work.

src/labelle/lib/devices/device_config.py
FaBjE: Will fix.

src/labelle/lib/devices/device_config.py
FaBjE: Thanks, will fix. (these are the kind of things you have no clue about when using a new language)

src/labelle/lib/devices/device_manager.py
FaBjE: ok, will fix

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Will fix

FaBjE: Good question...
Stupid answer: I learned about the raising exceptions later on. 

will fix.

src/labelle/lib/devices/device_manager.py
FaBjE: Yeah this was an ugly workaround. Will be obsolete when I fix the other function.
And yes of course I meant snake_case 😉

src/labelle/lib/devices/device_manager.py
FaBjE: I was not sure how to proceed than. Just print nothing?
Is there a quick syntax to "add if not null, otherwise leave empty"

src/labelle/lib/devices/device_config.py
FaBjE: Ok will fix

src/labelle/lib/devices/device_config.py
FaBjE: I want to make long lines, but there is a pre-commit hook from preventing me to...

I will add the extended description to the function comment.

src/labelle/lib/devices/device_config.py
FaBjE: If It makes you happy, I'll change it

src/labelle/lib/devices/device_config.py
FaBjE: The issue is that you need the tape with as input to calculate this. 
am I missing something?

src/labelle/lib/devices/device_config.py
FaBjE: The margins depend on the tape size selected as input.
A property that has a value based on another property's value seems very dirty to me.

I kept the bottom margin for the sake of completeness, they are equal now due to the calculation but that might not always be the case. The current rendering class only uses the top-margin but that could change too...

src/labelle/lib/devices/device_config.py
FaBjE: These are the result of the pre-commit hook that is configured. Maybe something is wrong on my end but I was not allowed to commit with these "long lines"

src/labelle/lib/devices/device_config.py
FaBjE: I'm OK with that, and while I'm add it, I'll try to fix it.
But keep in mind that I'm a python n00b. So my strategy is: don´t touch things, you might break it...

src/labelle/lib/devices/dymo_labeler.py
FaBjE: I want the first item in the tuple. Since it has no name how do you refer to it?

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Will fix.

src/labelle/lib/constants.py
FaBjE: I'll try (the append was just copied from an example)

maresb:
> I'm not happy with most of them. But it are either issues of me not knowing python syntax well enough or the linter/validator settings not allowing me to.

No worries! Please just do as well as you practically can, and we'll help out with the rest.

I agree that tape sizes are listed in mm, so for the UI it definitely makes a lot more sense than px. But personally I think of it more as a label. Like in my mental model, "12 mm" corresponds to the string `"12 mm"` rather than a numerical value `12.0` that can be computed with. As a result, doing a sanity check with mm to ensure the pixel counts are roughly correct is one thing, but it seems like a headache to get conversion and rounding to work out in a simultaneously simple and correct way.

> The LUT is a bit harder to generate for an inexperienced programmer than just measuring the size of your test-pattern print.

Ya, I think the process could be made simpler, like enhancing the test pattern by stuff like adding numbers to the larger blocks, and to print some additional lines and checkerboard patterns. But we're not generating these LUTs at scale, so I'd rather personally help the inexperienced users to get things dialed in precisely with units of pixels. Does that make sense?

Please don't worry about the rebasing, just do whatever's easiest for you.

Thanks a lot for taking such an ambitious approach here! If this gets to be too much, please let us know and we can think about how to break it apart into smaller features.

src/labelle/gui/q_settings_toolbar.py
maresb: Ah, tuples are "immutable" so you should always generate a new one instead of modifying an old one.

src/labelle/gui/gui.py
maresb: The `@property` thing is called a "decorator". That's a bit of a rabbit-hole...

Python's property mechanism is a way to spoof function calls to appear as attributes. They are popular, but also considered by some to be hacky.

I think it's fine for now if you just define a getter function.

src/labelle/lib/constants.py
maresb: Here's an example:

```python
def f(x, y):
    return 2 * x + y

f(y=1, x=2)  # 5
```

src/labelle/lib/devices/device_config.py
maresb:
```python
    """The inaccuracy of the tape position relative to the printhead

    Inaccuracy of the tape position is mostly caused by
    the tape moving slightly from side to side in the cartridge.
    Improper cartrigde placemement is also an factor,
    but negligible due to a physical spring in the lid.
    """
```

maresb: The purpose of the pre-commit hook is not to make things pretty. Instead the idea is to reduce entropy, so that there are fewer ways to express equivalent code. (Even if it's uglier, at least it's more standardized.)

If it gets in your way, feel free to ignore the errors for now and use `git commit --no-verify`, and we can deal with it once at the end.

maresb:
> I was thinking that if we encounter non-centered printers in the future

I thought we determined that your printer with the 9mm tape is non-centered with a top margin of 37 and a bottom margin of 27.

FaBjE:
> I agree that tape sizes are listed in mm, so for the UI it definitely makes a lot more sense than px. But personally I think of it more as a label. Like in my mental model, "12 mm" corresponds to the string `"12 mm"` rather than a numerical value `12.0` that can be computed with. As a result, doing a sanity check with mm to ensure the pixel counts are roughly correct is one thing, but it seems like a headache to get conversion and rounding to work out in a simultaneously simple and correct way.

True, but somewhere we must know the size of the label in pixels. I think we cannot assume that is a constant.
So it should be calculated based on the printer properties, I'll keep it as is for now. Feel free to update it somewhere in the future.

> > The LUT is a bit harder to generate for an inexperienced programmer than just measuring the size of your test-pattern print.
> 
> Ya, I think the process could be made simpler, like enhancing the test pattern by stuff like adding numbers to the larger blocks, and to print some additional lines and checkerboard patterns. But we're not generating these LUTs at scale, so I'd rather personally help the inexperienced users to get things dialed in precisely with units of pixels. Does that make sense?

Yes and no, I think it is more a thing of preference. 
But I honestly think the calculation is not that hard. At the moment we have the same precision we would have with pixels.

> Please don't worry about the rebasing, just do whatever's easiest for you.
> 
> Thanks a lot for taking such an ambitious approach here! If this gets to be too much, please let us know and we can think about how to break it apart into smaller features.
It's OK for now. But I don´t have all the time in the world. So My take is to get the code to an acceptable level to get it merged.
Than we can improve upon that "base" 

> > I was thinking that if we encounter non-centered printers in the future
> 
> I thought we determined in #11 that your printer with the 9mm tape is non-centered with a top margin of 37 and a bottom margin of 27.

Yes I thought so too, but than I calculated the following:
My printer is 128 pixels at 18mm. so 7.11p/mm
9 mm tape should be 63.99px, lets round to 64px.
Leaving 64 pixels as margin, for a centered tape this should be 32 at both sides.
37-32= 5, and 27-32=-5, so the tape has deviated 5 pixels from the center.
5 / 7.11 = 0.70mm. So the tape is 0.7mm off-center.
Playing with the tape cartridge a bit that would make sense as there is always a bit of play in the tape guides and position of the tape in the printer.

So in my opinion the 0.7mm (I rounded it to 1mm) is always the manufacturing tolerance we have to deal with.
You can calibrate it pixel-perfect on my printer. But there is no guarantee that somebody else with a (Same model) different printer and different tape is aligned the exact same way.

tomers:
Regarding `px` vs `mm` discussion, please refer to [Dymo's LW 550 Technical Reference Document](https://download.dymo.com/dymo/user-guides/LabelWriter/LW550Series/LW%20550%20Technical%20Reference.pdf), where they describe the properties of their print head (page 4):

> About the LabelWriter 550 Series Printers
The LabelWriter 550 series printers (LabelWriter 550, 550 Turbo, 5XL) are high-performance, low-cost
printers used for printing mailing labels, postage, file folder labels, barcode labels, and more. The
LabelWriter 550 and LabelWriter 550 Turbo printers have a 57 mm wide print head. The LabelWriter 5XL
has a 101 mm wide print head. All printers have 300-dpi print head.
The 57 mm wide thermal print head uses 672 individually addressable dots to form individual raster
lines of data at 300 dots per inch across the print head—the 101 mm wide print head uses 1248 dots.
Both use 300 dots per inch in the travel direction as directed by the print control data.

Note how they describe both the printer head width, the density (in `dpi`), and the number of pixels.
I think our inner configuration and logic should use pixels. User facing parts should be in millimeters.

tomek-szczesny:
It's not a veto but I don't agree that LUTs with two numerical parameters are more complicated than anything else.
Because that's what we ultimately need, a printout width in dots, and offset (bottom margin) in dots, for tapes smaller than a print head. Top margin isn't required.
All the rest about my LUTs may have been suboptimal because I suck at Python, but certainly I wouldn't bother amassing any more information than two dot values of our concern.

The formula that you brought up could be used by us internally to get estimated LUT values in case none of us has a printer or tape of interest.

Disclaimer: I am heavily biased towards simplicity and minimal logic, because most of my coding is for FPGAs. ;)

The test pattern was designed so it gives all the necessary information in one go. The idea was that an owner of an unsupported printer would share a test pattern photo, and we would update the LUT based on that. Ben said he could do that and certainly I could too. I also promised to write a manual before we all inevitably step down from our posts here.

Millimeters were a nightmare of the legacy code and I made it a point to remove them from everywhere except UI strings. Yes, that conversion constant wasn't precise anyway, and will not be even remotely true once we turn our attention to other thermal printers. At the very least, if we decide to stick to millimeters in cases like print margins settings and so on, these should be defined separately for each printer. 
Printers with different resolutions across x and y axes are so rare I wouldn't worry about that.

src/labelle/lib/constants.py
FaBjE: Fixed

src/labelle/lib/constants.py
FaBjE: Converted DeviceConfig to a dataclass

src/labelle/gui/gui.py
FaBjE: Fixed (I think)

maresb:
Thanks for the changes @FaBjE! This is shaping up very nicely.

> So the tape is 0.7mm off-center.
> Playing with the tape cartridge a bit that would make sense as there is always a bit of play in the tape guides and position of the tape in the printer.
>
> So in my opinion the 0.7mm (I rounded it to 1mm) is always the manufacturing tolerance we have to deal with.
> You can calibrate it pixel-perfect on my printer. But there is no guarantee that somebody else with a (Same model) different printer and different tape is aligned the exact same way.

It makes a lot more sense to me now, thanks for the context!

I'm disturbed by a discrepancy in the technical reference where they claim 300 dpi = 11.811 px/mm. But @FaBjE computes and verifies 7.11 px/mm. Any ideas what's going on here?

tomek-szczesny:
I measured my old test pattern from PnP, there are 48 dots across about 6.85mm -> 7.01/mm, or 178 dpi.

The most common thermal printer resolutions are 203dpi (8/mm), 300dpi and 600 dpi. I've never heard of thermal print heads below 200dpi.

My best guess is that they use >300dpi print head and do pixel doubling in firmware due to some problems found at the development stage.

FaBjE:
> Thanks for the changes @FaBjE! This is shaping up very nicely.

Not done yet 😉  But I won´t be able to finish everything today.

> > So the tape is 0.7mm off-center.
> > Playing with the tape cartridge a bit that would make sense as there is always a bit of play in the tape guides and position of the tape in the printer.
> > So in my opinion the 0.7mm (I rounded it to 1mm) is always the manufacturing tolerance we have to deal with.
> > You can calibrate it pixel-perfect on my printer. But there is no guarantee that somebody else with a (Same model) different printer and different tape is aligned the exact same way.
> 
> It makes a lot more sense to me now, thanks for the context!

No problem.

> I'm disturbed by a discrepancy in the technical reference where they claim 300 dpi = 11.811 px/mm. But @FaBjE computes and verifies 7.11 px/mm. Any ideas what's going on here?

The example with 300dpi is about the newer model printers. We have different older style printers. with about 180 dpi (180.6222)
 
> I will try to give the changes a more thorough inspection/test later today.

About that, do I click the "Resolve conversation" button, or do you do that when my changes are verified?

src/labelle/lib/devices/dymo_labeler.py
FaBjE: This is outdated btw

maresb:
> About that, do I click the "Resolve conversation" button, or do you do that when my changes are verified?

Ideally the person making the request (asking a question, requesting a change) resolves. If it's clear that you've decisively addressed the request and no followup is needed, then you can close yourself. It's also fine to leave them open. There are no rules, only common sense and etiquette.

src/labelle/lib/devices/device_config.py
FaBjE: (should be) fixed

src/labelle/lib/devices/device_config.py
FaBjE: Made dataclass

src/labelle/lib/devices/device_config.py
FaBjE: Outdated/fixed

src/labelle/lib/devices/device_config.py
FaBjE: Outdated

src/labelle/lib/constants.py
maresb: I did a search for backslashes, and all the files you changed are clean. Feel free to resolve this.

src/labelle/lib/devices/device_config.py
FaBjE: Changed

src/labelle/lib/devices/device_config.py
FaBjE: Moved to the labeler class and made it a property

src/labelle/lib/devices/device_config.py
FaBjE: Moved to the labeler class and made it a property / named tuple

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Made it a named tuple and used name

src/labelle/lib/devices/device_config.py
FaBjE: fixed

src/labelle/gui/q_settings_toolbar.py
FaBjE: Fixed (It didn´t complain this time)

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Fixed it, it now takes the highest supported size if not set

src/labelle/lib/devices/device_config.py
FaBjE: Unfortunately there is no document at this moment. I think we need to add a link when it is there.

src/labelle/lib/devices/device_config.py
FaBjE: Fixed

src/labelle/lib/devices/device_config.py
FaBjE: Fixed (now in dymo_labeler.py)

src/labelle/lib/devices/device_manager.py
FaBjE: Fixed

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Fixed

FaBjE: Fixed

src/labelle/lib/devices/device_manager.py
FaBjE: Fixed

src/labelle/lib/devices/device_manager.py
FaBjE: Fixed

src/labelle/lib/devices/device_config.py
FaBjE: Fixed as suggested.

src/labelle/lib/devices/device_config.py
FaBjE: Reworded / made it better aligned

src/labelle/lib/devices/device_config.py
FaBjE: Reworded / made it better aligned

src/labelle/lib/devices/device_config.py
FaBjE: Fixed

src/labelle/lib/devices/dymo_labeler.py
FaBjE: Fixed

FaBjE:
> > About that, do I click the "Resolve conversation" button, or do you do that when my changes are verified?
> 
> Ideally the person making the request (asking a question, requesting a change) resolves. If it's clear that you've decisively addressed the request and no followup is needed, then you can close yourself. It's also fine to leave them open. There are no rules, only common sense and etiquette.

Good to know. Makes sense to me, I've left most open as I think my fixed should be check too if they are the desired way.
I think I addressed all comments now.

FaBjE:
I did a big change for the DPI values. https://github.com/labelle-org/labelle/pull/54/commits/7246473e2914ffa7574c58155b7ca9147eba96a3
I'm not sure if I should make a separate PR for that, or include it here.

maresb (changes requested):
Thanks so much @FaBjE for your work here.

When I run `labelle-gui` with my LabelManager PnP, I get the following stack trace:

```python
qt.gui.icc: fromIccProfile: failed size sanity 2
Traceback (most recent call last):
  File "/home/mares/repos/labelle/src/labelle/gui/gui.py", line 108, in _on_settings_changed
    height_px=self._dymo_labeler.label_height_px,
  File "/home/mares/repos/labelle/src/labelle/lib/devices/dymo_labeler.py", line 277, in label_height_px
    return self.tape_print_properties.usable_tape_height_px
  File "/home/mares/repos/labelle/src/labelle/lib/devices/dymo_labeler.py", line 305, in tape_print_properties
    raise ValueError(
ValueError: Unsupported tape size 24mm. Supported sizes: [6, 9, 12]mm
Aborted (core dumped)
```

I believe this is indicative of a structural problem that we are introducing here. (Although the rest of the codebase is not innocent.) I strongly disapprove of the use of properties here. Accessing class attributes should be a completely safe operation. (The type checker should guarantee that they're defined.) In the stack trace we see two nested attribute accesses, which is an insane obfuscation. Unless there are exceptional circumstances, attributes should be static and functions should compute stuff. Note also how `tape_print_properties` is a huge block of (thankfully well-documented) code.

Instead of having a property called `label_height_px` I'd like to instead see an ordinary function `get_label_height_as_px`. The benefits to this approach are that:

1. we define an honest interface where attributes are attributes and functions are functions
2. it makes us actually think about what should be an attribute vs a function instead of lazily reencoding everything imaginable as a property, creating needless layers of indirection

But then this also brings up the issue of whether or not we should even have `mm` in the first case.

A tape has:
* height in mm
* height in px
* dpi / pixels_per_mm
and any two determines the third. As far as I can tell, we don't need DPI for anything, and it doesn't seem like we even need height in mm if we treat "tape size" as a label. While we can do some clever stuff with dpi for DYMO printers, I wonder if it's actually worth all this code complexity.

FaBjE:
> Thanks so much @FaBjE for your work here.

Ok, I'm going to be brutally honest here. Not that I want to, but I think it needs to happen.Up until this point I liked to contribute something to labelle and getting my label printer to work under linux.
But this is starting to change. 

The current codebase is/was not made for printers with different properties. (Supported tape sizes, dpi, etc) I tried to integrate this, but I needed to touch a lot of stuff. 

Trying to to change as little as possible (within reason) I did this. But there is a lot of stuff I think that can be handled different/better with the things we know now. Some may be personal preference or coding style preference. 
In my opinion, either we accept it is not as pretty as we want it, but it works. Or we do a big overhaul of the code and make everything pretty. 

But I think we are stuck in a no-mans-land between the two at this point.

As I said in the opening post of this PR this code is not perfect. There are issues with it, mostly caused by my changes "conflicting" the current setup of the codebase.

Let me address the points:

> When I run `labelle-gui` with my LabelManager PnP, I get the following stack trace:
> 
> ```python
> qt.gui.icc: fromIccProfile: failed size sanity 2
> Traceback (most recent call last):
>   File "/home/mares/repos/labelle/src/labelle/gui/gui.py", line 108, in _on_settings_changed
>     height_px=self._dymo_labeler.label_height_px,
>   File "/home/mares/repos/labelle/src/labelle/lib/devices/dymo_labeler.py", line 277, in label_height_px
>     return self.tape_print_properties.usable_tape_height_px
>   File "/home/mares/repos/labelle/src/labelle/lib/devices/dymo_labeler.py", line 305, in tape_print_properties
>     raise ValueError(
> ValueError: Unsupported tape size 24mm. Supported sizes: [6, 9, 12]mm
> Aborted (core dumped)
> ```
> 
> I believe this is indicative of a structural problem that we are introducing here. (Although the rest of the codebase is not innocent.) I strongly disapprove of the use of properties here. Accessing class attributes should be a completely safe operation. (The type checker should guarantee that they're defined.) In the stack trace we see two nested attribute accesses, which is an insane obfuscation. Unless there are exceptional circumstances, attributes should be static and functions should compute stuff. Note also how `tape_print_properties` is a huge block of (thankfully well-documented) code.
> 

Yes, this is caused by a structural problem the GUI has. For starters: I didn´t notice this as my printer supports all tapes.
But the GUI can´t (properly) handle changing printers (properties). 

More low level: when the printer properties are loaded (like supported tape sizes) the change-event of the control fires every time a item like a tape size is added/removed. 

It starts with the default (simulator) config. (which has all supported tape sizes) Than the USB printer is initialized and "attached" to the GUI. Than the weird stuff/problems starts to happen. The "backend" already has the USB printer. But the GUI controls still have the old data. Change events fire, and the GUI tries to set properties (left over from the default) that the printer doesn´t support. So boom, exception.

IMHO: The change events should not fire when the printer properties are being loaded.1. Disable change events, 2. Clear GUI controls, 3. Load new printer properties to controls, 4. Enable change events.

> Instead of having a property called `label_height_px` I'd like to instead see an ordinary function `get_label_height_as_px`. The benefits to this approach are that:
> 
>     1. we define an honest interface where attributes are attributes and functions are functions
>
>     2. it makes us actually think about what should be an attribute vs a function instead of lazily reencoding everything imaginable as a property, creating needless layers of indirection
> 

For starters, this `label_height_px` was always in the codebase. Originally it was named `height_px` as a property of the labeler object. Which is weird imho, because it is not the height of the labeler. (as I read it) but the tape height. So i did a rename.

Second, there already were a lot of properties that do either the same or something similar to eachother. 

Third, I was specifically asked to make this a property. https://github.com/labelle-org/labelle/pull/54#pullrequestreview-2133618583 If you are 100% sure you want it a function I will convert it back to a function, but that will be the last time I'm changing it.

> But then this also brings up the issue of whether or not we should even have `mm` in the first case.
> 
> A tape has:
> 
>     * height in mm

Correct

>     * height in px

No. The PRINTER has a height in px for a tape. There are no pixels on a tape.

>     * dpi / pixels_per_mm

No. The PRINTER has a DPI.

>       and any two determines the third. As far as I can tell, we don't need DPI for anything, and it doesn't seem like we even need height in mm if we treat "tape size" as a label. While we can do some clever stuff with dpi for DYMO printers, I wonder if it's actually worth all this code complexity.

You need DPI for the GUI (margins display) and CLI, calculations of tape size in pixels (for a particular printer) 
FYI: In my last commit the DPI from the CLI was set different than the DPI of the constants.

If I were to setup this code today, I would make the two of classes of 
1: Printer Config object, containing static data of a printer 
2: Printer runtime object, containing all functions of a printer and the "runtime" calculations. Of say for example the amount of pixels the printer can print on a (selected) tape size.

After that, pass the printer object to all classes that needs to to something. Like the renderers, GUI etc. They can retrieve the data they need from the printer object. 

They may need some more input, like for example a text or barcode that needs to be rendered. But all printer/tape specific data can just be retrieved from the printer object instead of passing around lots of separate variables.


Well, these are just my two cents. 
I´d like you to define a strategy, prospect, view or design or whatever you want to call it how you want to continue.
And I'll see if I want to be a part of that.

tomers:
> Third, I was specifically asked to make this a property. https://github.com/labelle-org/labelle/pull/54#pullrequestreview-2133618583 If you are 100% sure you want it a function I will convert it back to a function, but that will be the last time I'm changing it.

I suggested it, but I am not orthodox about it, and I do accept the reasoning raised earlier, as to why this was a bad suggestion. I am sorry for causing you to waste your eneregy, and lose some motivation to work on this codebase.
I hope you'll continue your excellent work, and keep improving it. The architecture indeed needs some thoughts and design in order to make it better. I love the idea of separating contant printer data (print head specification, etc.) and runtime data (current tape loaded).

maresb:
Hi @FaBjE, I'm really sorry that my criticism caused frustration and discouragement. I completely agree with everything you wrote. My criticism of the overuse of properties is more about the existing codebase and the need to change course rather than your contribution in particular. It was not so tactful for me to single out your code as an example of why property overuse leads to confusion. Especially since the previous review requested use of properties.

As you have realized, there are still a lot of structural issues in the codebase. A lot of things snuck in because I didn't review code as carefully as I should have. So it's a challenging situation. I want to be encouraging of contributions from people like you. But I also want the code to remain maintainable, and avoid further entrenching our previous bad design decisions.

You ask me to lay out a vision, but I think you already lay out a very clear vision in your last comment. The one thing I'd adjust is to avoid the idea of a monolithic printer object. I'd rather design things so that we have isolated stages, each of which only requires limited information. (Like the Linux philosophy of piping together simple commands.)

To figure out what these isolated stages should look like, let's suppose we were designing a web client as a replacement for the GUI. I think the first thing we'd need is a high-level specification of the label elements (e.g. text, barcode, etc.) that are to be horizontally combined. These shouldn't depend on the tape size or device, and it should ideally be easily serializable. Next would come the rasterization step that, for a given tape size in pixels, converts the elements into a bitmap to print. Finally, with the bitmap in hand, it should be passed to the printer for printing. So there are these steps: element specification, rasterization, printing to device, which I'd like to see as isolated as possible.

As for concrete next steps for this PR, let's try and make sure things are running without any errors. Let's try within reason to reduce the amount of stuff that's being added to interfaces and being passed around. And for these structural problems, let's keep track of what's wrong, and build in a way that will facilitate eventually fixing those structural problems. 

And I'll try and to better to keep things fun. I think we're all here to build cool stuff in a relaxing environment. Sorry if I was being too up-tight.

tomek-szczesny:
Since the project has a new name and a new scope (all label printers, ideally), it makes sense to define a sustainable architecture and overhaul it, thus bump the major release number. This might cost us less work in the long run, as with a few Dymos we already struggle to keep it tidy.

@maresb's general idea has some merit. I think that each printer model should be a different "object", with its own renderer code, attributes and so on.

One thing that worries me about this proposed approach is the second stage that composes the pieces together. I am thinking about barcodes and QR codes in particular, that will not work if scaled improperly. Thus this stage has to know how many dots it's working with, something that definitely is the domain of the third stage.

I guess we'll need a graph, and a space for this discussion :)

maresb:
> One thing that worries me about this proposed approach is the second stage that composes the pieces together. I am thinking about barcodes and QR codes in particular, that will not work if scaled improperly. Thus this stage has to know how many dots it's working with, something that definitely is the domain of the third stage.

Right, the pipeline can't be strictly one-way. In pseudocode I'd propose something like:

```python
def print_label(label_specification, device, tape):
    bitmap_height = device.usable_tape_height_px(tape)
    bitmap = rasterize(label_specification, bitmap_height)
    device.print(bitmap, tape)
```

Of course it probably will end up more complicated than this in practice due to implementation details.

So for this PR I think my specific question would be if, instead of passing around the `DymoLabeler` object, would it be sufficient and feasible to compute the bitmap height upfront and pass around only that instead? My hope is that this would be able to considerably simplify the code in this PR, and make it easier in the future to refactor. (It may well be that I'm wrong and that would require rewriting the whole codebase, but I hope not! :see_no_evil:)

FaBjE:
> I suggested it, but I am not orthodox about it, and I do accept the reasoning raised earlier, as to why this was a bad suggestion. I am sorry for causing you to waste your eneregy, and lose some motivation to work on this codebase. I hope you'll continue your excellent work, and keep improving it. The architecture indeed needs some thoughts and design in order to make it better. I love the idea of separating contant printer data (print head specification, etc.) and runtime data (current tape loaded).

I would not call it a bad suggestion. Because every other thing is put into a property in the dymo labeler class.
So in that way it made sense to me, and I changed it.


> Hi @FaBjE, I'm really sorry that my criticism caused frustration and discouragement. I completely agree with everything you wrote. My criticism of the overuse of properties is more about the existing codebase and the need to change course rather than your contribution in particular. It was not so tactful for me to single out your code as an example of why property overuse leads to confusion. Especially since the previous review requested use of properties.

Well, you didn´t exactly put my code on the spot as the propery already existed. (I only renamed it).
The confusing/frustrating part for me is that this PR is instead of discussing my changes is becoming more and more discussion about the global architecture of the project.
Something I had nothing to do with, and should not overhaul in this PR. 
I did the best I could with the current codebase. But as stated before, most things I encountered would not be my first choice (either). But made sense if you would set this up for 1 printer type.

> As you have realized, there are still a lot of structural issues in the codebase. A lot of things snuck in because I didn't review code as carefully as I should have. So it's a challenging situation. I want to be encouraging of contributions from people like you. But I also want the code to remain maintainable, and avoid further entrenching our previous bad design decisions.

Yes, but what you are currently doing now from my perspective is criticizing stuff out of scope of this PR.

> You ask me to lay out a vision, but I think you already lay out a very clear vision in your last comment. The one thing I'd adjust is to avoid the idea of a monolithic printer object. I'd rather design things so that we have isolated stages, each of which only requires limited information. (Like the Linux philosophy of piping together simple commands.)

Well, yes, but that is my vision. I have no clue in where you are standing on this or what the goal of the project is.

My advice on your view would be that it might be the most ideal design. (keeping stages separate) But be realistic and keep it simple. This is at the moment a tiny project with about no-one with time to code on it. You will introduce a abstraction and challenges with it. Which all require effort to code.

> To figure out what these isolated stages should look like, let's suppose we were designing a web client as a replacement for the GUI. I think the first thing we'd need is a high-level specification of the label elements (e.g. text, barcode, etc.) that are to be horizontally combined. These shouldn't depend on the tape size or device, and it should ideally be easily serializable. Next would come the rasterization step that, for a given tape size in pixels, converts the elements into a bitmap to print. Finally, with the bitmap in hand, it should be passed to the printer for printing. So there are these steps: element specification, rasterization, printing to device, which I'd like to see as isolated as possible.

Brutally honest, I (would like to) see labelle as the simple "simple but functional"/no-nonsense portable application that came on the Dymo PnP labeler (for windows) and I think you are 80% there atm.
Thats what I am looking for at least. Just printing some labels from a GUI. nothing fancy.

An example that I read in the GUI issue is about the office setting. But not to ruin your dreams or anything.
But the likely hood of something like labelle to be used in the foreseeable future are slim imho. For (professional) office settings there are much better (and expensive) solutions available.

I think labelle's strength will be in supporting the "older" hardware. and keeping it alive. reducing e-waste.
Even on windows the official dymo software is pretty much out of support for these types of printers.
But nothing is wrong with the label printers, I can buy tape cheaply at discount stores. and it can print labels fine.

Again, it may be fine ambitions, but something (too) far in the future imho.
Especially considering the amount of active coders on the project.

On the pratical side as @tomek-szczesny pointed out in it's next post, the bitmap that needs to be rendered is direct dependent on the printer (dpi) and label (size).

> As for concrete next steps for this PR, let's try and make sure things are running without any errors. Let's try within reason to reduce the amount of stuff that's being added to interfaces and being passed around. And for these structural problems, let's keep track of what's wrong, and build in a way that will facilitate eventually fixing those structural problems.
> 
> And I'll try and to better to keep things fun. I think we're all here to build cool stuff in a relaxing environment. Sorry if I was being too up-tight. I really hope that you stick around.

I have a different suggestion for this PR. I would like to see a "develop" branch.
For this branch we use this PR as starting point. 

A separate overhaul of the design may be implemented on top of that.

Maybe just in tiny steps to keep it manageable. (again, amount of available coders/time)
Whenever we are happy with it, the develop branch can be merged to the main branch.


> So for this PR I think my specific question would be if, instead of passing around the `DymoLabeler` object, would it be sufficient and feasible to compute the bitmap height upfront and pass around only that instead? My hope is that this would be able to considerably simplify the code in this PR, and make it easier in the future to refactor. (It may well be that I'm wrong and that would require rewriting the whole codebase, but I hope not! 🙈)

The dymo labeler object is current passed around for the DPI. Label height is already passed around.
there are lots more things to consider if you want to support more devices. (for example, label width, all newer types have fixed size labels instead of our "endless" tape).
But again, practically. This works. 

So again my view. Merge this to develop, agree on a global design, rework everything (in tiny steps), have a proper codebase.

maresb:
> So again my view. Merge this to develop, agree on a global design, rework everything (in tiny steps), have a proper codebase.

I'd be happy to do this. That way we can follow up separately with the structural issues I've raised before merging to `main`, people can build on that if they'd like, and we avoid further blocking on this PR. Great suggestion!

I just work on this project to screw around. I have no serious ambitions. For quite a while I was the only active maintainer, and I was just trying to keep the project alive. I do have things I'd like to eventually see implemented, like a web client and server, but unfortunately I don't have the free time to implement that stuff myself.

Also, to be honest I don't understand the details of how all these classes are implemented, and whenever I look at the details I'm surprised and confused. It used to be way worse until @tomers came along and rewrote major portions of the codebase. He took us from 20% to 80% of having a stable functional app.

maresb (approved):

PR #55, opened 2024-06-23
Improve sample pattern

maresb:
`--sample-pattern 64` on 9mm tape on LabelManager PnP:
image

`--sample-pattern 128` on 9mm tape on LabelManager PnP:
image
(This makes it easy (for me) to count backwards from the 80 and see that there is a faint row 65.)

* Horizontal lines now come in fours. (At least for me this makes it easier to count the four lines.)
* Horizontal lines are repeated both at the beginning and the end.
* The zero-indexed y-value of the last pixel in a box is printed when it fits.
* There is an extra width-4 stretch of 1×1 checkerboard.

To accommodate this, I rendered the test pattern as a `HorizontallyCombinedRenderEngine`. In order to fit the bitmaps into the `HorizontallyCombinedRenderEngine`, I had to wrap the `Image` class as a `_ImageRenderEngine` since the `PictureRenderEngine` takes a `Path` object instead of an `Image`.

CC also @FaBjE for review


tomek-szczesny:
So this is what it looks like when it is handled by capable hands! :)

I think that counting dots beginning with 1 would make counting easier, and the values more rounded. :)
I don't think this would cause any problems in the kitchen. After all the test print is to determine how many pixels are missing on each side. 

Other than that, perfect.

maresb:
@tomek-szczesny, good point! My new favorite Turing test is, "count to 10". :joy: 

image

FaBjE:
Fine with me 😅 
Maybe you can add little howto document so I can link to it from the comment (in the code)?

src/labelle/lib/render_engines/horizontally_combined.py
```diff
@@ -27,17 +29,24 @@ def render(self, context: RenderContext) -> Image.Image:
...
+def _compute_total_width(bitmaps: Sequence[Image.Image], padding: int) -> int:
+    number_of_bitmaps = len(bitmaps)
+    width_of_bitmaps = sum(b.width for b in bitmaps)
+
+    # We alternate bitmaps and padding, and have bitmaps on each end.
+    number_of_paddings = number_of_bitmaps - 1
+    width_of_paddings = number_of_paddings * padding
+
+    total_width = width_of_bitmaps + width_of_paddings
+    return total_width
```
tomers, line 52: I am not sure this decomposition of the equation to calculate total width is neccessary. It is too detailed.
You could just replace `self.PADDING` with `padding` in the original code.
I am fond of the idea to create intermediate `total_width` variable, since it helps readability, but to extract a one-liner to a function seems to me as an exaggaration.
Also, if you do extrac to a separate function, why not make is a part of the class?

tomers (approved):
By looking at the sample pattern, I still not sure I truly understand how to use it, i.e. figure out how many pixels are shown. Maybe a small markdown document will help...

Thanks for this contribution!
I think we should eventually support multiple sample patterns (ijn the CLI we can pass the pattern number)

src/labelle/lib/render_engines/horizontally_combined.py
maresb: I made it verbose because I don't ever want to make an effort to reconstruct the reason why I did `- 1`. Also, splitting it off as a function makes it easy to ignore the uninteresting detail, even if it could be done as a one-liner.

I don't see what the width computation has to do with the `RenderEngine` interface, which is why I didn't add it to the class.

maresb:
I need to make another set of photos after adding the `h=64` text.

maresb:
Partial draft of [a markdown doc](https://github.com/maresb/labelle/blob/improve-sample-pattern/doc/margin-calibration-howto.md)

doc/margin-calibration-howto.md
```diff
@@ -0,0 +1,44 @@
+# How to calibrate the label margins of DYMO printers
+
+Each model of DYMO printer has a stationary print head of a given width.
```
FaBjE, line 3: FYI: In the source it is referred to as height. (label height, etc)

doc/margin-calibration-howto.md
```diff
@@ -0,0 +1,44 @@
...
+Most DYMO printers support 6mm, 9mm, and 12mm D1 tapes
+and have a 64 pixel print head with a 9mm printable area.
+
+The LabelManager PC II also supports 19mm and 24mm D1 tapes
+and has a 128 pixel print head with a 19mm printable area.
```
FaBjE, line 8: 18 mm (instead of 19mm)

doc/margin-calibration-howto.md
```diff
@@ -0,0 +1,44 @@
...
+
```
FaBjE, line 9: I don't know if you want to list them all, but this page lists the tape options of several printers:
https://www.labelcity.com/dymo-d1-label-tape-compatibility-guide

Just mentioning that the PC II is not the only one to support 24mm

doc/margin-calibration-howto.md
FaBjE, line 1: Maybe add a little hands-on section with the commandline argument and what the user needs to do?
Like:
- Print label with labelle --test-pattern=64
- Look at missing pixels, note down first and last
- Print label with labelle --test-pattern=128
- bla bla bla
- Submit issue with folllowing pictures/data

FaBjE (changes requested):
Good start :) Added some tiny corrections and suggestions.

doc/margin-calibration-howto.md
tomek-szczesny: Actually all we need is for someone to print a test pattern once, with a sufficient number of dots.
We could ask for 256 dots and then to submitting the issue here, keep it simple.
In fact, 256 could be the default value of the --test-pattern switch.

maresb:
@FaBjE, thank you so much for your review, your comments are extremely helpful! I hope to get to it soon. (I'll also try and get to #51 soon.)

In order to be able to describe the general calibration procedure, would you be able to provide a few examples using this new test pattern with your 9mm tape? I'm thinking perhaps one test pattern with 128 px, and then another with 93 px? (My idea is that the top of your tape should be at pixel #91, so with 93 px there should be two rows cut off from the pattern on the top.)

FaBjE:
> @FaBjE, thank you so much for your review, your comments are extremely helpful! I hope to get to it soon. (I'll also try and get to #51 soon.)
> 
> In order to be able to describe the general calibration procedure, would you be able to provide a few examples using this new test pattern with your 9mm tape? I'm thinking perhaps one test pattern with 128 px, and then another with 93 px? (My idea is that the top of your tape should be at pixel #91, so with 93 px there should be two rows cut off from the pattern on the top.)

No worries. 

I have ordered 6mm and 19mm tape, they are in transit to me. I expect them in about a week.
Maybe for the sake of completeness I can print a series (128px test pattern on each tape size).
I can do a test-print with 93 px as well. Just send me a list of tape-size / pattern size you want me to test.

Practical issue though, the branch you are working on doesn't support my printer (yet 😉) 
I can maybe rebase it upon this branch of mine: https://github.com/labelle-org/labelle/compare/main...FaBjE:labelle:feature/addLabelManagerPCII That is just original codebase + my printer ID added.
It can´t print proper labels, but it should be OK for the test pattern I think?

maresb:
I just made another very quick pass on this. Still todo:

* [ ] Proper hands-on instructions as per https://github.com/labelle-org/labelle/pull/55#discussion_r1650387230
* [ ] Update the images
* [ ] Explain how to fix a dirty print head

I just produced a printout with some defects which I believe are due to dust on the print head, so I'll take the opportunity to explain cleaning.

maresb:
This is now much better thanks to your feedback @FaBjE! I have some structural changes in the works in order to bring `develop` back into `main`. If your tapes arrive, then it's probably easiest for now to create a branch merging this into `develop` so that you can print the updated pattern.

doc/margin-calibration-howto.md
```diff
@@ -0,0 +1,58 @@
+# How to calibrate the label margins of DYMO printers
+
+The goal of calibration is to determine:
+
+1. How many pixels does the print head have?
```
FaBjE, line 5: 3rd goal: Determine active area size of the printhead (mm)

maresb: What do you mean exactly? Do you mean that we should measure the height of the pattern in mm so that we can compute `px_per_mm`?

maresb:
Follow-up work for future PRs:

* Add cross-reference to the device specification in the source code once `develop` has been merged.
* Add example photo of test pattern for a tape that is significantly smaller than the print head size.

doc/margin-calibration-howto.md
FaBjE: Exactly. To make sense, this needs to be done on the largest supported tape/label size.

maresb:
Merging and releasing this so that we can use it for #62

PR #66, opened 2024-08-03
Develop cleanup 1

maresb:
Some simple cleanups for the develop branch, more to come.

### DIFF
--- a/src/labelle/lib/devices/device_config.py
+++ b/src/labelle/lib/devices/device_config.py
@@ -35,9 +35,8 @@ class DeviceConfig:
     but negligible due to a physical spring in the lid.
     """
 
-    LABELER_DISTANCE_BETWEEN_PRINT_HEAD_AND_CUTTER_MM: float = 8.1
-    """This is the distance between the printhead and the cutter (knife).
-    """
+    distance_between_print_head_and_cutter_px: int = 57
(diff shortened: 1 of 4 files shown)


src/labelle/lib/devices/device_config.py
tomers, line 1: Why is this commit needed, can you please elaborate?
I assume that when adding a new device, one measures the printhead cutter distance using a ruler. How can one measure the distance in pixels?
I think this is a physical value, thus it should be in a real distance unit.

maresb: > I assume that when adding a new device, one measures the printhead cutter distance using a ruler. How can one measure the distance in pixels?

Hmm, my assumption was that we should measure the distance using a test pattern. I hadn't thought about using a ruler.

In the end we must advance the tape by some whole number of pixels, so at least in my mind pixels are the fundamental unit of length.

maresb: > I assume that when adding a new device, one measures the printhead cutter distance using a ruler. How can one measure the distance in pixels?

Hmm, my assumption was that we should measure the distance using a test pattern. I hadn't thought about using a ruler.

In the end we must advance the tape by some whole number of pixels, so at least in my mind pixels are the fundamental unit of length.

PR #71, opened 2024-08-08
Document USB issues in README.md

### LINKED ISSUE #68: Frequent overflow issues
Hello, and THANK YOU for Labelle!

I have a LabelManager 280 on Linux and using 12mm label tape.

I have found that when the text size exceeds approximately 50mm in width, I get frequent "overflow" errors and only part of the label prints.

Attempting to print after that is likely to yield further errors, including possible I/O errors, until I either power cycle the 280 or unplug it from USB and plug it back in again.

This is on Debian Linux and I would be happy to provide whatever debugging would be helpful!

jgoerzen:
Fixes #68


README.md
```diff
@@ -73,6 +73,13 @@ or on Arch with
 sudo pacman -S python-pipx
 ```
 
+You will also need the USB development libraries; on Ubuntu/Debian, install
+with
+
+```bash
+sudo apt-get  --no-install-recommends install libusb1-dev
```
maresb, line 80: Why is `libusb1-dev` necessary? What error message do you get without it? Would `libusb-1.0-0` suffice?

README.md
```diff
@@ -90,6 +97,14 @@ You do not have sufficient access to the device. You probably want to add the a
 
 <!-- markdownlint-enable MD013 -->
 
+## Interference from CUPS
```
maresb, line 100: I thought you were still having issues after disabling CUPS: https://github.com/labelle-org/labelle/issues/68#issuecomment-2273356623

Also any ideas why you're the first to report this? Could it be specific to the 280?

README.md
```diff
@@ -90,6 +97,14 @@ You do not have sufficient access to the device. You probably want to add the a
...
+
+Sometimes, the CUPS printing system will attempt to communicate with the label
+maker, causing interference with Labelle.  If you don't otherwise have a USB
```
maresb, line 103: Not necessary, but do you have any ideas for if someone does have a USB printer? For example, is it possible to make this rule only apply to a specific USB device?

README.md
```diff
@@ -90,6 +97,14 @@ You do not have sufficient access to the device. You probably want to add the a
...
+printer, you can disable CUPS attempting this with:
+
+sudo touch /etc/udev/rules.d/70-printers.rules
```
maresb, line 106: Could you please add a fenced block around this command?

maresb (commented):
Thanks @jgoerzen for putting this together!

README.md
jgoerzen: The error was that "No supported devices found."  That is present even with the library but without the -dev package.  It appears that one of the vendored libraries tries to compile something against it or something.

README.md
jgoerzen: Very occasionally.  Before disabling CUPS, labelle was nearly unusable.  Now I occasionally see that, let's call it maybe 2% of the time instead of being easily reproducible almost every time.  I believe it to be a different issue.

README.md
jgoerzen: I made an effort at that, but couldn't figure out how to block the system udev rules that way.  There may be a way but I couldn't find it.

README.md
jgoerzen: Yes, will push shortly.

README.md
jgoerzen: Actually, you are correct.  I made a mistake with my testing.  I will fix this.

maresb: Great, thanks so much for tracking this down!

Would you be able to explicitly mention the "No supported devices found" error? Something like,

> If your printer is connected but you receive the error `No supported devices found` then please ensure `libusb` has been installed. On Ubuntu/Debian it can be installed with

README.md
maresb: Similarly, could you please explicitly mention the overflow error in relation to this workaround?

jgoerzen: Done in 2dd0b48

README.md
maresb: Sorry @jgoerzen, did you see my above comment?

README.md
maresb: Thanks!

README.md
maresb: Resolved in #77.

maresb:
I merged part of this in #77. For the rest, let's see how #76 works out.

maresb:
Thanks again @jgoerzen for digging into this and for the PR.

Re-reading the #68 thread, I think CUPS was a red herring for the overflow, and I think #76 addresses the actual cause, so the CUPS workaround shouldn't be needed anymore.

I'm going to close this PR as superseded by #76. If anyone does run into genuine CUPS device contention (independent of the overflow) after #76, please reopen or file a fresh issue and we'll document it properly. Thanks again for the careful investigation.

PR #76, opened 2024-09-28
Fix Frequent overflow issues #68 on LabelManager 280

poita66:
The overflow happens because the LabelManager 280 sends more than 8 bytes and the buffer is too small. This needs testing on other devices, but I expect that it'll be fine for them too.

LibUSB has no way of knowing how much data the device is going to send so if the supplied buffer size is too small, the overflow happens.

See https://libusb.sourceforge.io/api-1.0/libusb_packetoverflow.html for more info

### DIFF
--- a/src/labelle/lib/devices/dymo_labeler.py
+++ b/src/labelle/lib/devices/dymo_labeler.py
@@ -98,7 +98,10 @@ class DymoLabelerFunctions:
                 # Send a status request
                 cmdBin = array.array("B", [ESC, ord("A")])
                 cmdBin.tofile(self._devout)
-                rspBin = self._devin.read(8)
+                # Increase buffer size to 16 bytes to prevent overflow
+                # on LabelManager 280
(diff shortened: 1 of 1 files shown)


maresb:
Amazing, thanks @poita66! A brief test shows that this seems to work on the LabelManager PnP.

I don't know anything about USB buffer sizes. What is the tradeoff for increasing this to 16? In particular, why stop at 16? Would anything go wrong if we increased to 512? (512 seems to work for me as well.)

@jgoerzen, does this help for your situation?

@tomek-szczesny, @tomers, any other devices we could test this with?

poita66:
@maresb You make a great point. I'll test and update the PR. I only doubled it as my other projects are embedded and I was in the RAM-conservative mindset.

tomek-szczesny:
Seems to be working fine on PnP. :)

maresb:
I like this. I'm going to assume this is good until someone complains. Let's cut a release.

Issue #78, opened 2024-09-29
--fixed-length Flag Doesn't Work Alone


poita66:
Example:

```$ python3 -m labelle.cli.cli --fixed-length 100 --justify center "CLAMPS"                                                           
Usage: python -m labelle.cli.cli [OPTIONS] [TEXT]... COMMAND [ARGS]...
Try 'python -m labelle.cli.cli --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value: Cannot specify min/max and fixed length at the same time                                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

poita66:
Fix incoming

PR #79, opened 2024-09-29
Fix --fixed-length Flag Doesn't Work Alone

poita66:
Fixes https://github.com/labelle-org/labelle/issues/78

Using the --fixed-length flag currently yields the following error:

```$ python3 -m labelle.cli.cli --fixed-length 100 --justify center "CLAMPS"                                                           
Usage: python -m labelle.cli.cli [OPTIONS] [TEXT]... COMMAND [ARGS]...
Try 'python -m labelle.cli.cli --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Invalid value: Cannot specify min/max and fixed length at the same time                                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

This PR allows the --min-length flag to be 0 or None

### DIFF
--- a/src/labelle/cli/cli.py
+++ b/src/labelle/cli/cli.py
@@ -444,8 +444,7 @@ def default(
             "Cannot specify both barcode with text and regular barcode"
         )
 
-    if fixed_length is not None and (
-        not (min_length is None or min_length == 0) or max_length is not None
+    if fixed_length is not None and min_length is not None max_length is not None
     ):
         raise typer.BadParameter(
             "Cannot specify min/max and fixed length at the same time"
--- a/src/labelle/cli/cli.py
+++ b/src/labelle/cli/cli.py
@@ -444,8 +444,7 @@ def default(
             "Cannot specify both barcode with text and regular barcode"
         )
 
-    if fixed_length is not None and min_length is not None max_length is not None
-    ):
+    if fixed_length is not None and min_length is not None max_length is not None):
         raise typer.BadParameter(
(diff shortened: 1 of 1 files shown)


maresb:
Thanks for catching this! Can't we get rid of the min_length == 0 check?

poita66:
That's what I originally did, but thought that we should retain the existing ability to provide both --min-length and --fixed-length in case anyone was already using them together. This way it's not a breaking change.

maresb:
Ah ya, I see your point. But in this case I think it was a pure bug. Semantically it doesn't make sense to have `--min-length=0` and `--fixed-length=x`. I'd rather just fix it and consider it as a "bugfix" rather than a "breaking change". Given the error message I think it should be straightforward for anyone who used the `--min-length=0` to figure out, but honestly you're probably the first (or at worst second) person to hit this.

poita66:
Yeah, fair. I'll update the PR

poita66:
Sorry for the mess of commits. I did them on my phone, which was a mistake. I'll squash them on my laptop in the morning.

maresb:
No worries and no rush! Thanks so much for the contribution!

maresb:
This looks good to me, so I'm just going to squash-merge it. Thanks!!!

PR #81, opened 2024-09-29
Support for LabelManager Wireless PnP

thinkl33t:
Some initial work on LabelManager Wireless PnP

- add auto cut to end of print statement (this is added to every print - in testing with a LabelManager PnP this doesn't effect anything
- add 24mm to valid lines

still to do:

- printer doesn't output anything when cold-booted until it has been printed to once by another application - currently i'm using [dymoconf/render-label.py](https://github.com/simon-budig/dymoconf/blob/master/render-label.py) and sending it to port 9100 on the printer to do this initialization
- when things are printed out, they come out very squished - i *suspect* i need to fiddle with `LABELER_PRINT_HEAD_HEIGHT_MM` depending on the specific labeller but i could be wrong here - i can't find any codepaths that differ between devices at the moment

### DIFF
--- a/src/labelle/lib/devices/dymo_labeler.py
+++ b/src/labelle/lib/devices/dymo_labeler.py
@@ -222,6 +222,8 @@ class DymoLabelerFunctions:
             self._raw_print_label(lines[0 : self._maxLines])
             del lines[0 : self._maxLines]
         self._raw_print_label(lines)
+        self._cut()
+        self._send_command()
 
     def _raw_print_label(self, lines: list[list[int]]):
         """Print the label described by lines (HLF)."""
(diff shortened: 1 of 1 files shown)


maresb:
The current situation with the codebase is unfortunately a bit precarious. We have forked into a `develop` branch which is unfortunately broken, but that adds some stuff to sanely support other devices. There's still a bit of cleanup to do before we can bring it into main though, so I'd recommend keeping this PR slim to avoid duplicating work and increasing the likelihood that we can merge.

maresb:
I appreciate a lot that you're showing your work early. For purely organizational purposes I'm going to convert this to a draft. Please feel free to mark it as "ready for review" and ping us whenever you'd like us to take another look at it.

thinkl33t:
After some more experimentation, sending the cut for every printer seems to crash out printers with no cutters, so i'm going to try and rework this against your `develop` branch :)

maresb:
That's strange, which printers are you testing against? I'm fairly confident that with my PnP the cut command was a noop, but I don't have it at the moment.

Please be warned that there's no timeline for merging `develop` into `main`, just to calibrate your expectations.

willumpie82:
Picking this up per @maresb's suggestion above to target `develop` — opened #156, which builds directly on the auto-cut and 24mm-support work here, plus validates this device's `DeviceConfig` (the placeholder values were the source of the "squished" output) and fixes a couple of protocol-level bugs found along the way. Thanks for the groundwork, @thinkl33t!

PR #82, opened 2024-09-29
Develop cleanup 2

maresb:
Following up from https://github.com/labelle-org/labelle/pull/66

### DIFF
--- a/src/labelle/lib/render_engines/margins.py
+++ b/src/labelle/lib/render_engines/margins.py
@@ -1,17 +1,18 @@
 from __future__ import annotations
 
-from typing import Literal, NamedTuple
+from typing import Literal
 
 from PIL import Image
 
 from labelle.lib.constants import Direction
 from labelle.lib.env_config import is_dev_mode_no_margins
+from labelle.lib.margins import LabelMarginsPx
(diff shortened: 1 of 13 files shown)


src/labelle/lib/render_engines/margins.py
```diff
@@ -14,49 +13,64 @@
 )
 
 
+class LabelMarginsPx(NamedTuple):
+    horizontal: int
+    vertical: int
+
```
tomers: According to Python docs, `NamedTuple` type hint is a "Typed version of [collections.namedtuple()](https://docs.python.org/3/library/collections.html#collections.namedtuple)". See https://docs.python.org/3/library/typing.html#typing.NamedTuple

Don't you prefer to use `collections.namedtuple`, instead of a regular Python class?

maresb: Good question. Some benefits:

1. Types. With `collections.namedtuple` you just have names, without the types, so you wouldn't get type checking.
2. Syntax. This behaves and looks like a normal Python class, but the init and string representation are managed for you. This also makes it easy to switch out with similar data containers, like `pydantic.BaseModel` in case you need input validation.

PR #101, opened 2025-02-06
Add barcode to cli --batch

da-Kai:
Added a `BARCODE` option to the --batch CLI parameter, allowing users to specify the type of barcode to generate.


maresb:
Thanks @da-Kai! This seems to me like a reasonable proposal. The batch format still needs to be fleshed out, so this is a great step in that direction.

Note that this introduces (in the single `BARCODE` case) optional arguments to commands via a `#` character separator before the first `:` character.

@jgoerzen created the batch format in https://github.com/labelle-org/labelle/pull/72. @jgoerzen, do you have any thoughts on this extension?

jgoerzen:
I like it - nice work!

maresb (approved):
@da-Kai, the type annotation `line: str` belongs on line 537 (at its first definition) rather than line 547. If you fix this then the mypy the check will pass. You can fix the formatting by making a comment here with the text "pre-commit.ci autofix"

da-Kai:
pre-commit.ci autofix

maresb:
Thanks @da-Kai!

PR #115, opened 2025-07-19
Add support for Letratag 200B

Gusser93:
This should add support for the Letratag 200B Bluetooth Low Energy Label Maker. For this to work, I made an abstract class `Device` to support USB and BLE devices. I also put the logic to send the commands in the Device classes instead of the LabelMaker class to allow different workflows for different connections.
- [ ] I still want to figure out if it is possible to generate the data for the image with PIL instead of numpy.
- [ ] I have to test the CLI tool
- [ ] I intend to implement some default values depending on the connected device. Currently, the Text on the Letratag seems a bit small and the margins too large.
- [ ] Someone has to test if the USB connection still works. I don't have access to these devices.
- [ ] A trigger to scan for BLE devices is still missing. Currently, this happens only when the application is started.
- [ ] No name is shown, as put the logic for that in the wrong place.
- [ ] I have to improve error handling for BLE devices.
- [ ] I have to test which commands are supported by the Letratag 200B and document these
- [ ] I still have to rewrite the instruct methods


maresb:
Looks very exciting, thanks @Gusser93!!!

> I still want to figure out if it is possible to generate the data for the image with PIL instead of numpy.

That would be really great!

I'm also wondering how to handle the bleak dependency. Maybe make it optional? Maybe we should also make the PyUSB dependency optional? :thinking:

PR #137, opened 2026-01-11
Make Labelle compatible with python-barcode v0.16

### LINKED ISSUE #123: python-barcode breaking change
python-barcode made a breaking change in it's API but versionned this as a minor rather than a major. 

The signature of the [BaseWriter](https://github.com/WhyNotHugo/python-barcode/blob/f013b69dc63874924a2b88b90e5028be3e7bdc11/barcode/writer.py#L114) no longer has default values. 

Reverting back to 0.15.1 does the trick. I don't think I have the rights to make a PR but this is all that is needed while you investigate further
```
diff --git a/pyproject.toml b/pyproject.toml
index 1074abd..b7d32ac 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "platformdirs",
     "Pillow>=8.1.2,<11",
     "PyQRCode>=1.2.1,<2",
-    "python-barcode>=0.13.1,<1",
+    "python-barcode>=0.13.1,<0.16",
     "pyusb",
     "PyQt6",
     "darkdetect",
```

Otherwise, labelle cannot print barcodes

claui:
In [python-barcode](https://pypi.org/project/python-barcode/) v0.16.0, the API of the BaseWriter constructor was [narrowed](https://github.com/WhyNotHugo/python-barcode/compare/v0.15.1...v0.16.0#diff-01c9de34453d29913aa28bb39e4c847e2ecf95cf81b1264c2759090f870152c9) so two of its callbacks can no longer be None (see issue #123.)

Given we override the implementation that would use those callbacks anyway, this PR passes dummy callbacks. This ensures compatibility with both pre-v0.16 and newer versions of python-barcode.

Additionally, loosen one of the assertions where the upstream wording of an error message slightly changed.

Fixes #123.

/cc @blaisegarant, @maresb

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     "platformdirs",
     "Pillow>=8.1.2,<11",
     "PyQRCode>=1.2.1,<2",
-    "python-barcode>=0.13.1,<0.16",
+    "python-barcode>=0.13.1",
     "pyusb",
     "PyQt6",
     "darkdetect",
--- a/src/labelle/lib/barcode_writer.py
+++ b/src/labelle/lib/barcode_writer.py
@@ -14,6 +14,10 @@ BinaryString = NewType("BinaryString", str)
 """A string that's been validated to contain only '0's and '1's."""
 
 
+def _noop() -> None:
+    pass
+
+
 def _validate_string_as_binary(s: str) -> BinaryString:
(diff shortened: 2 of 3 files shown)


maresb:
Thanks @claui for tracking this down!

